PR #25, opened 2015-01-13
Fixed broken gemspec

kirs:
`.gemtest` doesn't exist:

```
$ gem build benchmark-ips.gemspec
WARNING:  See http://guides.rubygems.org/specification-reference/ for help
ERROR:  While executing gem ... (Gem::InvalidSpecificationException)
    [".gemtest"] are not files
```


evanphx:
You don't need to build it. I should just remove that rake task. the gemspec picks up the right version automatically.

PR #26, opened 2015-01-15
Accept symbol as report's argument.

JuanitoFatas:
Now we could pass symbol to `#report` like the following :blush: :

```ruby
require 'benchmark/ips'

data = [*0..100_000_000]

Benchmark.ips do |x|
  x.report(:find)    { data.find    { |number| number > 77_777_777 } }
  x.report(:bsearch) { data.bsearch { |number| number > 77_777_777 } }
end
```


jemc:
:+1:
Just ran into this and was surprised it didn't work.

JuanitoFatas:
:beers:

PR #34, opened 2015-03-10
Allow passing a custom suite via config

splattael:
For ROM we'd like to disable GC during benchmarks: https://github.com/rom-rb/rom/blob/9ad85095d5fe1b15fd7880899995cd4aadc95197/benchmarks/setup.rb#L56,#L91

After discovering `suite` in `Benchmark::IPS` I wondered if I could pass a suite instance easier.
Defining a `Benchmark::Suite` class which "magically" will be used feels strange to me.

With this PR I'm proposing to pass a suite instance via config like:

```ruby
require 'benchmark/ips'

class GCSuite
  def warming(*)
    run_gc
  end

  def running(*)
    run_gc
  end

  def warmup_stats(*)
  end

  def add_report(*)
  end

  private

  def run_gc
    GC.enable
    GC.start
    GC.disable
  end
end

suite = GCSuite.new 

Benchmark.ips do |x|
  x.config(:suite => suite)
  x.report("job1") { ... }
  x.report("job2") { ... }
end
```

*Update: Suite exmaple completed*

### DIFF
--- a/lib/benchmark/ips/job.rb
+++ b/lib/benchmark/ips/job.rb
@@ -142,6 +142,7 @@ module Benchmark
       def config opts
         @warmup = opts[:warmup] if opts[:warmup]
         @time = opts[:time] if opts[:time]
+        @suite = opts[:suite] if opts[:suite]
       end
 
       # Return true if job needs to be compared.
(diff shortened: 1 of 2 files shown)


splattael:
Thanks for merging! :)

PR #36, opened 2015-03-13
Quiet mode

kirs:
With the `quite` keyword argument we can disable the output.

This feature is quite needed for [ruby-bench](https://github.com/ruby-bench/) and Rails: https://github.com/ruby-bench/ruby-bench-suite/pull/8/files#diff-9c2f5a1dd03f27a99f4330ee96d77d1fR81


kirs:
If everything is OK with the idea, I will add some tests. /cc @evanphx

evanphx:
Please use "quiet=false", not the ruby 2.2 only syntax.

kirs:
@evanphx done! Also tests added.

PR #38, opened 2015-03-19
Support keyword arguments

kirs:
I think `Benchmark.ips(time: 10.seconds, warmup: true, quiet: true)` should be better readable.
@evanphx


evanphx:
No problem with the idea but can you please change the test to not use 1.9 only syntax? This gem still supports 1.8.

evanphx:
@kirs Any update on this?

PR #40, opened 2015-07-11
Remove ree and 1.8.7 from travis

PragTob:
* they make travis fail and are not supported anymore anyway

I just made it a PR because it's so easy - your call of course (could also be allowed failures) but I think their life time has ended for such a long time so it's cool to get rid of them.


kbrock:
It isn't that hard to add 1.8.7 support, see #40.

But agreed, lets get travis passing.

PR #41, opened 2015-07-19
Support keyword arguments (1.8 friendly)

kbrock:
This is basically #38 but with ruby 1.8 syntax
Good idea

Fixes #38

I can rebase after you merge #40 or #42 - it needs those to pass


evanphx:
Go ahead and rebase this.

evanphx:
Thanks for all the great work @kbrock

PR #43, opened 2015-07-19
allow any datatype for labels

kbrock:
When paired with json, this allows more rich metadata.

I'm having trouble coming up with descriptive labels that I want to use.
So I'm passing in a hash. the hash is being passed through to the json output


evanphx:
Wanna rebase this against master to pickup the json fixes too?

kbrock:
@evanphx thanks so much

PR #44, opened 2015-07-21
move job entry into own file

kbrock:
Hi @evanphx 

Do you have any thoughts about having Job::Entry in its own file?

just close if you'd prefer to keep them in the same file.

thnx


evanphx:
Seems fine to me!

PR #46, opened 2015-07-26
Report stdout

kbrock:
The suite and the output to standard out were very similar.

Added a "suite" that printed to standard out.

- moved headers into report. It is not standard to the suite API. Can move around if you have thoughts.
- used tell don't ask for json and compare output
- can rename stdout report, just give me a name.


evanphx:
`StdoutReport` seems like a fine name.

PR #48, opened 2015-08-11
Fix runtime comparison

bquorning:
I kept getting an

    undefined method `rjust' for
    #<Benchmark::IPS::Report::Entry:0x007ff53b8a2950> (NoMethodError)

Makes me wonder if this code is even used by anyone?

### DIFF
--- a/lib/benchmark/compare.rb
+++ b/lib/benchmark/compare.rb
@@ -49,7 +49,7 @@ module Benchmark
       if iter
         $stdout.printf "%20s: %10.1f i/s\n", best.label, best.ips
       else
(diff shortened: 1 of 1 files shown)


bquorning:
Any comment on whether this code is even reachable? I imagine someone would have reported a bug earlier if it were.

evanphx:
Ok, tracking it all the way back, it looks like that code path is just rarely if ever used. It provides the ability to compare reports that don't do `ips`, but no one uses it for anything else. We could probably just remove that other code path at this point.

bquorning:
I may open a PR for removing that code path, then :-)

PR #51, opened 2015-08-12
All reports respond to ips

bquorning:
As mentioned in #48, it looks like half the code in lib/benchmark/compare.rb isn’t used at all. Doesn’t all reports do `:ips`?

cc @kbrock who changed the code recently.


kbrock:
:+1: good call

PR #55, opened 2016-02-12
Separate messages for warming up and calculating

chrisseaton:
They both just say 'Calculating' at the moment - this patch changes them to say 'Warming up' or 'Sampling'.


evanphx:
I'd prefer to have the second way still say "Calculating..." because users are expecting that. No reason to change that label.

chrisseaton:
I changed it to say 'Warming up' and 'Calculating'.

PR #56, opened 2016-02-12
Add support for hold! and independent invocations.

chrisseaton:
benchmark-ips is often used to compare different ways to implement the same functionality, but running two benchmarks in the same invocation of Ruby is not going to tell you whether A or B is faster, it's going to tell you if A, or *B-given-that-A-has-already-run* is faster.

For the theory on why this is not ideal, see research such as Kalibera and Jones [1].

For a practical example of why this is not ideal I can illustrate with this benchmark:

```ruby
require 'benchmark/ips'

class Foo
  def method_under_test
    10
  end
end

class Bar
  def method_under_test
    20
  end
end

def call_method_under_test(x)
  x.method_under_test
end

Benchmark.ips do |x|
  
  foo = Foo.new

  x.report("test_a") do
    call_method_under_test foo
  end
  
  bar = Bar.new
  
  x.report("test_b") do
    call_method_under_test bar
  end

  # x.hold! - I'll explain this later on 
  x.compare!
  
end
```

The problem is that in an optimising implementation of Ruby will compile the call `x.method_under_test` as monomorphic for the first benchmark, and bimorphic for the second. The second benchmark is disadvantaged and the results we get for the two benchmarks are dependent on their order.

I can demonstrate this with JRuby+Truffle. I run the benchmark, asking it to only compile the method `call_method_under_test` and turning off splitting and OSR to reduce the noise (otherwise understanding what is being compiled and why is hard).

```
$ JAVACMD=../../graal/GraalVM-0.9/jre/bin/javao bin/jruby -X+T -J-server -J-G:+TraceTruffleCompilation -J-G:TruffleCompileOnly=call_method_under_test -J-G:-TruffleOSR -J-G:-TruffleSplitting -I ../benchmark-ips/lib dependence.rb
```

We see `call_method_under_test` compiled twice, once for each benchmark (it has to be recompiled, as the compiler compiled it with the assumption that it would always be monomorphic baked in, but in some cases the first benchmark keeps the original monomorphic copy and only the second gets the bimorphic one). The first copy is 164 bytes of code, the second is 188. The second benchmark is larger and slower because it came second and so is polymorphic.

```
[truffle] opt done         call_method_under_test:dependence.rb:15 <opt>               |ASTSize      14/   18 |Time   287( 259+28  )ms |DirectCallNodes I    1/D    0 |GraalNodes    35/   28 |CodeSize          164 |Source dependence.rb:15 
[truffle] opt done         call_method_under_test:dependence.rb:15 <opt>               |ASTSize      14/   22 |Time   155( 145+10  )ms |DirectCallNodes I    2/D    0 |GraalNodes    40/   25 |CodeSize          188 |Source dependence.rb:15 
```

My solution to this is to run each benchmark in an independent VM. I added a method `hold!` (like `compare!`) that runs one benchmark, saves the results, and prompts you to run Ruby again for the next benchmark.

```
$ ... bin/jruby -X+T ...
...
[truffle] opt done         call_method_under_test:dependence.rb:15 <opt>               |ASTSize      14/   18 |Time   252( 214+39  )ms |DirectCallNodes I    1/D    0 |GraalNodes    35/   28 |CodeSize          164 |Source dependence.rb:15
...
Pausing here -- run Ruby again to measure the next benchmark...
$ ... bin/jruby -X+T ...
...
[truffle] opt done         call_method_under_test:dependence.rb:15 <opt>               |ASTSize      14/   18 |Time   252( 214+39  )ms |DirectCallNodes I    1/D    0 |GraalNodes    35/   28 |CodeSize          164 |Source dependence.rb:15 
...
```

Now I get two methods, both 164 bytes, and I've so made the benchmarks independent (or a bit more independent).

This is optional - you only need to use it if you want to. It saves data in a little file that is checksummed so it's invalidated if you benchmark changes.

[1] https://kar.kent.ac.uk/33611/7/paper.pdf

### DIFF
--- a/lib/benchmark/ips/job.rb
+++ b/lib/benchmark/ips/job.rb
@@ -18,6 +18,10 @@ module Benchmark
       # @return [Boolean] true if needs to run compare.
       attr_reader :compare
 
+      # Determining whether to hold results between Ruby invocations
+      # @return [Boolean]
+      attr_accessor :hold
+
       # Report object containing information about the run.
       # @return [Report] the report object.
       attr_reader :full_report
@@ -34,6 +38,10 @@ module Benchmark
       # @return [Integer]
       attr_accessor :time
 
+      # The hash of the benchmark source.
+      # @return [String]
+      attr_accessor :source_hash
+
       # Instantiate the Benchmark::IPS::Job.
       # @option opts [Benchmark::Suite] (nil) :suite Specify Benchmark::Suite.
       # @option opts [Boolean] (false) :quiet Suppress the printing of information.
@@ -43,6 +51,7 @@ module Benchmark
         @list = []
         @compare = false
         @json_path = false
+        @source_hash = nil
 
         @timing = {}
         @full_report = Report.new
@@ -72,6 +81,16 @@ module Benchmark
         @compare = true
       end
 
+      # Return true if results are held while multiple Ruby invocations
+      # @return [Boolean] Need to hold results between multiple Ruby invocations?
+      def hold?
+        @hold
+      end
+
+      # Set @hold to true.
+      def hold!
+        @hold = true
+      end
 
       # Return true if job needs to generate json.
       # @return [Boolean] Need to generate json?
@@ -135,6 +154,8 @@ module Benchmark
       def run_warmup
         @stdout.start_warming if @stdout
         @list.each do |item|
+          next if hold? && item.held_results?(self)
+          
           @suite.warming item.label, @warmup if @suite
           @stdout.warming item.label, @warmup if @stdout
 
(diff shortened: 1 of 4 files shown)


evanphx:
I don't think Proc#source_location is available in 1.8, so this would break 1.8.

I understand the desire and I think I'd prefer instead of magically inferring a name, a name should have to passed to `#hold!` that is the path to store the held data in.

chrisseaton:
Updated to use a file specified as `hold! 'file.data'`. The simplified the implementation a bit as well, actually.

lib/benchmark/ips/job.rb
```diff
@@ -34,6 +38,10 @@ class Job
       # @return [Integer]
       attr_accessor :time
 
+      # The hash of the benchmark source.
+      # @return [String]
+      attr_accessor :source_hash
```
evanphx: You can remove this too, not necessary with the explicit file name.

chrisseaton: Sorry - yes fixed.

PR #62, opened 2016-04-01
use the monotonic clock for getting the current time

tenderlove:
This has a few advantages over using Time.now.  First the monotonic
clock will not change even if the system clock changes while the
benchmark is running.  Second, we don't have to coerce Time objects to
floats every time we want to calculate the test runtime.  Third, we
don't have to allocate Time objects on each iteration which /should/
reduce stddev noise from the GC.

### DIFF
--- a/lib/benchmark/ips/job.rb
+++ b/lib/benchmark/ips/job.rb
@@ -5,7 +5,7 @@ module Benchmark
       # Microseconds per 100 millisecond.
       MICROSECONDS_PER_100MS = 100_000
       # Microseconds per second.
-      MICROSECONDS_PER_SECOND = 1_000_000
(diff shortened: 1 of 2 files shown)


kbrock:
I keep reading the title as "use the **atomic** clock for getting the current time"

@evanphx Those test failures are all bundle failures. Should we drop support for <1.8.7 and 1.9.3?

evanphx:
Using `Time.now.to_f * 1_000` was far, far too slow to use in a hot loop it seemed. I reworked the logic for https://github.com/evanphx/benchmark-ips/commit/39ac3095590a203d00b86a7f54931e9228b34873.

tenderlove:
@evanphx one other thing I was thinking is that we could use nanoseconds and avoid floating point math (though I'm not sure if it matters)

PR #65, opened 2016-04-04
Fixup inject

kbrock:
Just minor tweaks to the syntax.

They can probably be all ignored
If there are any in particular you don't want, just say the word

### DIFF
--- a/lib/benchmark/ips/job.rb
+++ b/lib/benchmark/ips/job.rb
@@ -259,7 +259,7 @@ module Benchmark
 
           final_time = before
 
-          measured_us = measurements_us.inject(0) { |a,i| a + i }
+          measured_us = measurements_us.inject(:+)
 
           all_ips = measurements_us.map { |time_us|
             iterations_per_sec cycles, time_us
--- a/lib/benchmark/ips/job.rb
+++ b/lib/benchmark/ips/job.rb
@@ -133,8 +133,7 @@ module Benchmark
       # @return [Integer] Cycles per 100ms.
       def cycles_per_100ms time_msec, iters
         cycles = ((MICROSECONDS_PER_100MS / time_msec) * iters).to_i
-        cycles = 1 if cycles <= 0
-        cycles
+        cycles <= 0 ? 1 : cycles
       end
 
       # Calculate the time difference of before and after in microseconds.
--- a/lib/benchmark/timing.rb
+++ b/lib/benchmark/timing.rb
@@ -35,14 +35,9 @@ module Benchmark
     # @param [Integer] resample_times Resample times, defaults to 100.
     # @return [Array] Resampled samples.
     def self.resample_mean(samples, resample_times=100)
-      resamples = []
-
-      resample_times.times do
-        resample = samples.map { samples[rand(samples.size)] }
-        resamples << Timing.mean(resample)
-      end
-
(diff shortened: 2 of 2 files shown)


lib/benchmark/ips/job.rb
```diff
@@ -260,7 +259,7 @@ def run_benchmark
 
           final_time = Timing.now
 
-          measured_us = measurements_us.inject(0) { |a,i| a + i }
+          measured_us = measurements_us.inject(&:+)
```
evanphx: I'm almost positive this change is a lot slower than the old way.

kbrock: I remembered something about that as well. But looks like for all but 1.8.7, the symbol form is 10% faster https://github.com/JuanitoFatas/fast-ruby/pull/105

Here is what I found:

```bash
[master] fast-ruby/code $ chruby 1.8.7
[master] fast-ruby/code $ ruby enumerable/inject-sum-vs-block.rb 
[...]
Comparison:
          inject sum:        0.1 i/s
        inject block:        0.0 i/s - 1.74x slower
[master] fast-ruby/code $ chruby 1.9.3
[master] fast-ruby/code $ ruby enumerable/inject-sum-vs-block.rb 
Comparison:
        inject block:        1.6 i/s
          inject sum:        1.5 i/s - 1.10x slower
[master] fast-ruby/code $ chruby 2.0.0
[master] fast-ruby/code $ ruby enumerable/inject-sum-vs-block.rb 
Comparison:
        inject block:        1.5 i/s
          inject sum:        1.3 i/s - 1.20x slower
[master] fast-ruby/code $ chruby 2.1.5
[master] fast-ruby/code $ ruby enumerable/inject-sum-vs-block.rb 
[...]
Comparison:
        inject block:        1.5 i/s
          inject sum:        1.3 i/s - 1.15x slower
[master] fast-ruby/code $ chruby 2.2.4
[master] fast-ruby/code $ ruby enumerable/inject-sum-vs-block.rb 
[...]
Comparison:
        inject block:        1.6 i/s
          inject sum:        1.4 i/s - 1.12x slower

File: lib/benchmark/timing.rb [hunk 2]
```diff
@@ -35,14 +35,9 @@ def self.stddev(samples, m=nil)
     # @param [Integer] resample_times Resample times, defaults to 100.
     # @return [Array] Resampled samples.
     def self.resample_mean(samples, resample_times=100)
-      resamples = []
-
-      resample_times.times do
-        resample = samples.map { samples[rand(samples.size)] }
-        resamples << Timing.mean(resample)
+      resample_times.times.map do
+        Timing.mean(samples.map { samples[rand(samples.size)] })
```
zenspider: Please use `{}` here to follow Weirich method

kbrock:
@zenspider changed `map` to use `{}` per Jim Weirich even though it spans multiple lines.

I tried on a single line and even though it wasn't too many characters, it was just too dense for me.

zenspider:
I don't see any reason why the CI failed. It borks on bundle against json only on the 1.8 flavors. @evanphx ? Any clue? 

I also can't figure out how to simply restart the CI and have it give it another go.

evanphx:
Weird. Looks fine otherwise, I'll see about CI.

PR #69, opened 2016-05-16
Add support for confidence intervals

chrisseaton:
Current output:

```
                 mul   243.455k i/100ms
                 pow   274.906k i/100ms
                 mul      5.635M (± 8.4%) i/s -     27.997M in   5.014078s
                 pow      7.674M (± 5.1%) i/s -     38.487M in   5.029545s

Comparison:
                 pow:  7674355.1 i/s
                 mul:  5635018.0 i/s - 1.36x  slower
```

With confidence intervals:

```
                 mul   244.458k i/100ms
                 pow   282.247k i/100ms
                 mul      5.996M (± 0.6%) i/s -     30.068M in   5.020666s
                 pow      8.035M (± 1.1%) i/s -     39.797M in   5.000516s
                   with 95.0% confidence

Comparison:
                 pow:  8034594.6 i/s
                 mul:  5996097.6 i/s - 1.34x  (± 0.02) slower
                   with 95.0% confidence
```

Why are confidence intervals good? The standard deviation isn't really actionable. If I tell you something is plus/mins X SD, what can you do with that? If I tell you something is plus/minus X and I'm 95% confident about that then you can theoretically use that in a quantitive assessment of the risk and cost of being wrong and use that to make a decision. It also isn't parametric - you can't make it smaller if you want more certainty, or larger if you are more relaxed.

Another big benefit is that we can show a confidence interval for the comparison as well! This isn't something that isn't possible at the moment.

Finally, I think the standard deviation is overly conservative, and confidence intervals are smaller in practice. From experience using `benchmark-ips`, the standard deviations we currently use are not useful because they're so large.

Adds an optional dependency on the `kalibera` gem.

@thedarkone what do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,14 +193,23 @@ can be calculated instead. A bootstrap confidence interval has the advantages of
 arguably being more mathematically sound for this application than a standard
 deviation, it additionally produces an error for relative slowdowns, which the
 standard deviation does not, and it is arguably more intuitive and actionable.
-W
-hen a bootstrap confidence interval is used, a median of the interval is used
+
+When a bootstrap confidence interval is used, a median of the interval is used
 rather than the mean of the samples, which is what you get with the default
 standard deviation.
+
+The bootstrap confidence interval used is the one described by Tomas Kalibera.
+Note that for this technique to be valid your benchmark should have reached a
+non-periodic steady state with statistically independent samples (it should
+have warmed up) by the time measurements start.
 
 Using a bootstrap confidence internal requires that the 'kalibera' gem is
 installed separately. This gem is not a formal dependency, as by default it is
 not needed.
+
+```
+gem install kalibera
+```
 
 ```ruby
 Benchmark.ips do |x|
--- a/README.md
+++ b/README.md
@@ -185,6 +185,38 @@ with `SHARE=1` argument. I.e.: `SHARE=1 ruby my_benchmark.rb`.
 Result will be sent to [benchmark.fyi](https://benchmark.fyi/) and benchmark-ips
 will display the link to share the benchmark's result.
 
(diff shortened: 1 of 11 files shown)


thedarkone:
> @thedarkone what do you think?

Haha, you make my PR #68 look like something coming from a village idiot 😆.

This is how it handles my bench from #68:

```
 work(1000 + r(500))    15.000  i/100ms
 work(1000 + r(400))    16.000  i/100ms
 work(1000 + r(500))    155.906  (± 0.9%) i/s -    780.000  in   5.008610s
 work(1000 + r(400))    161.551  (± 0.7%) i/s -    816.000  in   5.053923s
                   with 95.0% confidence

Comparison:
 work(1000 + r(400)):      161.6 i/s
 work(1000 + r(500)):      155.9 i/s - 1.04x  (± 0.01) slower
                   with 95.0% confidence
```

Awesome!

chrisseaton:
Here's some slides from the Benchmarking '16 conference about the confidence intervals I'm using http://soft-dev.org/events/bench16/slides/Tomas_Kalibera.pdf

chrisseaton:
And this is what confidence intervals look like compared to the SD for the ERB benchmark from MRI.

erb

chrisseaton:
@evanphx it looks like this is now failing CI because master is (I just merged).

Beside that, do you have any opinions on the PR?

The key advantage is it gives you a CI for the *speedup* as well as the absolute measurement.

evanphx:
Looks great! I think the keys changes can end up on disk but those are saved briefly so it shouldn't be an issue. 

I'll check CI in the morning, but looks fine!

kbrock:
Thanks @chrisseaton - this is great

PR #72, opened 2016-08-06
Add missing files to Manifest

jonatack:
Stack trace:

    active_support/dependencies.rb:293:in `require': cannot load such file -- benchmark/ips/stats/sd (LoadError)

    gems/ruby-2.3.1/gems/benchmark-ips-2.7.0/lib/benchmark/ips.rb:4:in `<top (required)>'


evanphx:
Fixed elsewhere.

jonatack:
You're welcome 🙄

PR #88, opened 2018-08-28
[WIP] Extensible jobs

kbrock:
Hi @evanphx 

I find I often want to extend `benchmark-ips` to add number of objects allocated or number of queries performed and others.

There are a few memory alternatives like https://github.com/michaelherold/benchmark-memory (there was another, can't find it), but to be honest, it would be nice if there were just one base for this mostly common code. I was hoping to keep `benchmark-ips` as the base glue that would be able to run multiple types of benchmarks. (probably not your main goal)

I think the first 2 commits are no-brainers.
Possibly the 3rd is as well.
If you prefer, I can pull some of those out into a separate PR and go from there.

### DIFF
--- a/lib/benchmark/compare.rb
+++ b/lib/benchmark/compare.rb
@@ -26,12 +26,17 @@ module Benchmark
   #   Reduce using to_proc:   247295.4 i/s - 1.13x slower
   #
   # Besides regular Calculating report, this will also indicates which one is slower.
-  module Compare
+  class Compare
+
+    # currently the core of the run is part of job#run and job#run_benchmark
+    def run(job)
+    end
 
     # Compare between reports, prints out facts of each report:
     # runtime, comparative speed difference.
-    # @param entries [Array<Report::Entry>] Reports to compare.
-    def compare(*entries)
+    # @param entries [Benchmark::IPS::Report] Report to compare.
+    def post_run(full_report)
+      entries = full_report.entries
       return if entries.size < 2
 
       sorted = entries.sort_by{ |e| e.stats.central_tendency }.reverse
@@ -67,6 +72,4 @@ module Benchmark
       $stdout.puts
     end
   end
-
-  extend Benchmark::Compare
 end
--- a/lib/benchmark/ips.rb
+++ b/lib/benchmark/ips.rb
@@ -1,6 +1,5 @@
 # encoding: utf-8
 require 'benchmark/timing'
-require 'benchmark/compare'
 require 'benchmark/ips/stats/stats_metric'
 require 'benchmark/ips/stats/sd'
 require 'benchmark/ips/stats/bootstrap'
--- a/lib/benchmark/ips.rb
+++ b/lib/benchmark/ips.rb
@@ -55,7 +55,7 @@ module Benchmark
 
       yield job
 
-      job.load_held_results
+      job.pre_run
 
       job.run
 
@@ -67,18 +67,9 @@ module Benchmark
       end
 
       $stdout.sync = sync
-      job.run_comparison
-      job.generate_json
+      job.post_run
 
-      report = job.full_report
-
-      if ENV['SHARE'] || ENV['SHARE_URL']
-        require 'benchmark/ips/share'
-        share = Share.new report, job
-        share.share
-      end
-
-      report
+      job.full_report
     end
 
     # Set options for running the benchmarks.
--- a/lib/benchmark/ips/job.rb
+++ b/lib/benchmark/ips/job.rb
@@ -220,6 +220,21 @@ module Benchmark
         File.delete @held_path if File.exist?(@held_path)
       end
 
+      def pre_run
+        load_held_results if hold? && held_results?
+      end
+
+      def post_run
+        run_comparison
+        generate_json
+
+        if ENV['SHARE'] || ENV['SHARE_URL']
+          require 'benchmark/ips/share'
+          share = Share.new full_report, self
+          share.share
+        end
+      end
+
       def run
         if @warmup && @warmup != 0 then
           @stdout.start_warming if @stdout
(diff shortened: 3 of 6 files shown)


evanphx:
This is great! Really no qualms about adding it. Do you have an memory one maybe we could add to the release too?

kbrock:
I do wondering about the overall concept of hold.

You can run 2 kinds of benchmarks (probably more but...)

A. 1 file with a bunch of implementations of the same method. e.g.: [fast-ruby](https://github.com/JuanitoFatas/fast-ruby)
B. 1 file with a bunch of implementations of different methods. e.g.: [ruby bench](https://github.com/ruby-bench/ruby-bench)

It seems that `hold` is for case B. But it becomes very manual. Improving a method and then running with 2 implementations seems to add quite a few steps. May be nice to run each version of the library from a CI.

Does it makes more sense to just use an api (i.e.: benchmark.fyi) for running across multiple implementations? This is essentially what the `hold.csv` file is. Right? common data across multiple runs. it can be a save file or a process that has an API.

We would need more blob/json based stats to store memory, queries, timings.
We would need meta-data on the implementation that was run for this particular test. e.g.: ruby version, library version, time run.
But besides that, it is very similar to benchmark.fyi

Think I'm getting ahead of myself though.
And maybe I'm just reinventing ruby-bench.

evanphx:
The hold system could definitely be better. It was originally design to help benchmark ruby implementation changes, that's why it saves and requires you to run it again.

kbrock:
@evanphx Hmm. This is a great conversation. But I'm sorry I conflated these 2 conversation here. I'll open another issue or pr with some thoughts.

@michaelherold I would like to define the examples once, and then use them across both / other test types. I had thought this could be extensible through the `Suite` interface. But realized the `Suite` is more focused on what happens between tests, and this is more focused on what happens in an individual test.

michaelherold:
@evanphx It does that job very well -- that's what I use it for, in fact. 😄

kbrock:
ok, pushed up some changes.
I'm glad you asked me to extend this, because I found a few issues in the implementation.

1. reworked more of this than I had expected.
2. Not sure if the helper interface should be `run(full_report)` or `run(job)`.
3. Not sure if you want me to add `Benchmark::IPS::Objects`. Seems work from @michaelherold is more thorough. Hope it will be a simple example if nothing else.

```ruby
require "benchmark/ips"

# comparison helper
# current interface passes a report object to run (since all seem to use full_report.entries or full_report.data)
# we need job.list, so we pass it as part of the initializer (could change the interface to run)
class CompareObjects
  def initialize(job)
    @job = job
  end

  # Compare between reports, prints out facts of each report:
  # objects allocated
  # @param report [Benchmark::IPS::Report] Report with entries to compare.
  def run(_report)
    entries = @job.list
    return if entries.size < 2

    objects_key = RUBY_VERSION < '2.2' ? :total_allocated_object : :total_allocated_objects

    allocations = {}

    entries.each do |entry|
      before = GC.stat[objects_key]
      entry.call_times(1)
      after = GC.stat[objects_key]
      allocations[entry] = after - before
    end

    sorted = entries.sort_by{ |e| allocations[e] }.reverse

    best = sorted.shift

    $stdout.puts "\nComparison:"
    $stdout.printf "%20s: %10d objects\n", best.label, allocations[best]

    sorted.each do |report|
      name = report.label.to_s
      $stdout.printf "%20s: %10d\n", name, allocations[report]
      $stdout.puts
    end

    $stdout.puts
  end
end

# under test

class Def
  def initialized
    @defined = true
  end

  # yes, I'm intentionally setting a different variable
  # want it to remain undefined

  # this is the winner (they are all very close)
  def undefqc # question colon 
    defined?(@undefined) ? @undefined : @defined = true
  end

  def defqc
    defined?(@defined) ? @defined : @defined = true
  end
end

d = Def.new

# benchmark

Benchmark.ips(warmup: 1, time: 1) do |x|
  x.report("undef ?:") { d.undefqc }
  x.report("def ?:") { d.defqc }
  #x.compare!
  x.add_helper(:objects, CompareObjects.new(x))
end
```

```
ruby -Ilib cast_def_microbenchmark.rb 
warmup: Benchmark::IPS::Job::Entry (true)
            undef ?:   267.706k i/100ms
warmup: Benchmark::IPS::Job::Entry (true)
              def ?:   273.221k i/100ms
            undef ?:      8.716M (± 2.9%) i/s -      8.834M in   1.014423s
              def ?:      8.878M (± 3.4%) i/s -      9.016M in   1.016769s

Comparison:
              def ?::          1 objects
            undef ?::          1
```

kbrock:
ok - I'm going back in.

I'll put this into a memory report - hoping michael can tear my implementation out and make it as great as his gem

I feel the data should go into the `Report::Entry` rather than just printing it out

kbrock:
```ruby
require "benchmark/ips"
require "active_record"

ActiveRecord::Base.establish_connection(ENV.fetch('DATABASE_URL') { "postgres://localhost/user_benchmark" })
ActiveRecord::Migration.verbose = false

ActiveRecord::Schema.define do
  create_table :users, force: true do |t|
    t.string :name
  end
  #add_index :users, :name
end
class User < ActiveRecord::Base ;end

if User.count == 0
  puts "Creating 100 users"
  100.times { |i| User.create name: "user #{i}" }
end

Benchmark.ips do |x|
  x.report("User.all.first") { User.all.first }
  x.report("User.all.to_a.first") { User.all.to_a.first }

  x.compare!
  x.queries!
  x.objects!
end
```

```
Creating 100 users
      User.all.first   287.000  i/100ms
 User.all.to_a.first    81.000  i/100ms
      User.all.first      2.962k (±14.5%) i/s -     14.637k in   5.081636s
 User.all.to_a.first    955.136  (± 7.2%) i/s -      4.779k in   5.034635s
Queries:
      User.all.first:   1 queries     1 objects
 User.all.to_a.first:   1 queries   100 objects

Comparison:
      User.all.first:     2962.3 i/s
 User.all.to_a.first:      955.1 i/s - 3.10x  slower

Queries Comparison:
 User.all.to_a.first:   1 queries   100 objects
      User.all.first:   1 queries     1 objects


Comparison:
 User.all.to_a.first:       1184 objects
      User.all.first:        148 objects
```

evanphx:
Do you want to close this one and continue to split it up?

kbrock:
ok, I'll do that. thanks (and sorry for time delay)

PR #90, opened 2018-10-27
Save job (alternative to hold)

kbrock:
Hello @evanphx and @michaelherold 

This takes the `hold!` conversation out of #88

When I am benchmarking, I tend to view the benchmark script as a single file, not as a unique test per invocation. I screw up the order dependencies and stuff.

I think I like running the whole script at a single pass, but with unique labels.

```bash
> chruby ruby-2.3.7 ; SAVE_FILE=run1 ruby benchmark_split.rb
> chruby ruby-2.4.4 ; SAVE_FILE=run1 ruby benchmark_split.rb
```

```ruby
require 'benchmark/ips'
require 'active_support/all'

NSTRING = nil
DELIMITER='/'.freeze
STRING="ab/cd/ef/gh".freeze

Benchmark.ips(:quiet => true) do |x|
  x.report({version: RUBY_VERSION, data: nil, method: "to_s.split"},  "NSTRING.to_s.split(DELIMITER)")
  x.report({version: RUBY_VERSION, data: nil, method: "?split:[]"},   "NSTRING ? NSTRING.split(DELIMITER) : []")
  x.report({version: RUBY_VERSION, data: nil, method: "&&split||[]"}, "NSTRING && NSTRING.split(DELIMITER) || []")
if RUBY_VERSION >= "2.4"
  x.report({version: RUBY_VERSION, data: nil, method: "&.split||[]"}, "NSTRING&.split(DELIMITER) || []")
end
  x.compare!
  x.save! "#{ENV['SAVE_FILE']}-nil.out" if ENV['SAVE_FILE']
end

Benchmark.ips(:quiet => true) do |x|
  x.report({version: RUBY_VERSION, data: :str, method: "to_s.split"},  "STRING.to_s.split(DELIMITER)")
  x.report({version: RUBY_VERSION, data: :str, method: "?split:[]"},   "STRING ? STRING.split(DELIMITER) : []")
  x.report({version: RUBY_VERSION, data: :str, method: "&&split||[]"}, "STRING && STRING.split(DELIMITER) || []")
if RUBY_VERSION >= "2.4"
  x.report({version: RUBY_VERSION, data: :str, method: "&.split||[]"}, "STRING&.split(DELIMITER) || []")
end
  x.compare!
  x.save! "#{ENV['SAVE_FILE']}-nil.out" if ENV['SAVE_FILE']
end
```

Granted I still need to stitch this stuff back together again, but that is another problem and another PR for another day.

Sorry I changed stats, some rubydoc, and some extra blank lines while I was in there.

Thanks,
Keenan

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,11 @@ This will run only one benchmarks each time you run the command, storing
 results in the specified file. The file is deleted when all results have been
 gathered and the report is shown.
 
+Alternatively, if you prefer a different approach, the `save!` command is
+available. Examples for [hold!](examples/hold.rb) and [save!](examples/save.rb) are available in
+the `examples/` directory.
+
+
 ### Multiple iterations
 
 In some cases you may want to run multiple iterations of the warmup and
--- a/examples/save.rb
+++ b/examples/save.rb
@@ -1,0 +1,50 @@
+#!/usr/bin/env ruby
+
+# example to explain save!
+# The save! feature expects to be run twice, generally with different Rubys.
+# save! can also be used to compare modules changes which impact the run time
+#
+# If you're comparing ruby versions, Just use the version in the label
+#
+#     x.report("ruby #{RUBY_VERSION}") { 'Bruce'.inspect }
+#
+# Or use a hash
+# 
+#     x.report("version" => RUBY_VERSION, "method" => 'bruce') { 'Bruce'.inspect }
+#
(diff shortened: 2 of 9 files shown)


kbrock:
@evanphx if you don't like the save changes, would you be willing to accept the stats changes in?

I am leveraging a bit of this code for `benchmark-sweet` which is a less invasive way to use a few suites together. Working locally but still in alpha.

evanphx:
Merged! Looks good!

PR #93, opened 2019-09-07
Add Ruby 2.3+ to travis

ixti:
FWIW I believe Ruby < 2.3 support can be dropped, as 2.3 is EOL and 2.4
will be EOL soon.


ixti:
Would be awesome to have gem release too (so that we can use `#save!` feature)

PR #98, opened 2020-04-24
Ignore empty saved files

jcoleman:
The current code fails if a save file is specified and already exists on disk but is empty. That makes it harder to script using temp files, since you'd generally want to pre-create the temp file.

While it's possible to workaround this by writing `"[]"` to the file ahead of time, you essentially need to read the code to realize that's why you're getting a "NoMethodError each on nil" exception, so it seems better to just handle this up front.

### DIFF
--- a/lib/benchmark/ips/job.rb
+++ b/lib/benchmark/ips/job.rb
@@ -187,7 +187,7 @@ module Benchmark
       end
 
       def load_held_results
-        return unless @held_path && File.exist?(@held_path)
+        return unless @held_path && !File.zero?(@held_path)
         require "json"
         @held_results = {}
         JSON.load(IO.read(@held_path)).each do |result|


jcoleman:
Looks like CI is broken generally:

```
$ rvm use 1.9.2 --install --binary --fuzzy
181curl: (22) The requested URL returned error: 404 Not Found
182Required ruby-1.9.2 is not installed - installing.
183curl: (22) The requested URL returned error: 404 Not Found
184Searching for binary rubies, this might take some time.
185Requested binary installation but no rubies are available to download, consider skipping --binary flag.
186Gemset '' does not exist, 'rvm ruby-1.9.2 do rvm gemset create ' first, or append '--create'.
187The command "rvm use 1.9.2 --install --binary --fuzzy" failed and exited with 2 during .
```

lib/benchmark/ips/job.rb
```diff
@@ -187,10 +187,10 @@ def iterations_per_sec cycles, time_us
       end
 
       def load_held_results
-        return unless @held_path && File.exist?(@held_path)
+        return unless @held_path && File.exist?(@held_path) && !(contents = IO.read(@held_path)).empty?
         require "json"
         @held_results = {}
-        JSON.load(IO.read(@held_path)).each do |result|
+        JSON.load(contents).each do |result|
```
nateberkopec, line 193: I don't see how the `contents` variable is necessary.

nateberkopec:
CI should be better now, you should rebase

lib/benchmark/ips/job.rb
jcoleman: How would you do this without it? We need to be able to read the file once, check the state of that read, and then use the contents of that read in parsing. That would seem to require a variable?

nateberkopec: Switch the File.exist? check to `File.zero?` https://github.com/ruby/ruby/blob/a0c7c23c9cec0d0ffcba012279cd652d28ad5bf3/file.c#L1995

nateberkopec: er, `!File.zero?` ofc

jcoleman: That means we hit the filesystem twice unnecessarily, and it also represents a race condition.

nateberkopec: IMO the cost of hitting the filesystem twice instead of once in a method called once per benchmark job is not important. I'd rather keep a two-element conditional and avoid a variable assignment to keep the complexity down.

I also don't understand the race condition?

jcoleman: The race condition is inherent in hitting the filesystem twice: the file might not exist and then exist (but be empty) at the read.

I realize that's very much a theoretical concern in the sense that it's very unlikely, but I tend to have a preference for writing it as theoretically correct given the complexity is quite minor.

I can change it if that's your preference (I'd rather get it merged obviously), but I did want to at least explain my thought process.

nateberkopec: I see. Given the current implementation of load_held_results, clear_held_results, and save_held_results, I would say that benchmark-ips does not support using a held results file that may be written to (or in this case, emptied) by another process.

Sorry for pushing you so hard on a 2-line change. Thanks for explaining your thinking + I appreciate your patience.

eregon: An alternative to make this more readable would be to do something like this:
```ruby
        return unless @held_path && File.exist?(@held_path)
        contents = IO.read(@held_path)
        return if contents.empty?
```

eregon: The `File.zero?` approach seems fine too to me, and is quite a bit simpler indeed.

nateberkopec:
Thanks @jcoleman !

PR #107, opened 2020-12-03
Fix broken hold! when results file didn't exist

ivoanjo:
In ed5c005071, the fix to allow the held results to be empty broke use
of `hold!` when no pre-existing file existed.

This is because `File.zero?` of a non-existing file returns `false`,
and thus in `load_held_results` the early `return` would not happen,
resulting in a failed attempt to parse the non-existing file:

```
$ ruby examples/hold.rb
Traceback (most recent call last):
	3: from examples/hold.rb:22:in `<main>'
	2: from ~/ruby/benchmark-ips/lib/benchmark/ips.rb:58:in `ips'
	1: from ~/ruby/benchmark-ips/lib/benchmark/ips/job.rb:193:in `load_held_results'
~/ruby/benchmark-ips/lib/benchmark/ips/job.rb:193:in `read': No such file or directory @ rb_sysopen - temp_results (Errno::ENOENT)
```

the full fix here is to check that both:

* there is a file
* and it is not empty

### DIFF
--- a/lib/benchmark/ips/job.rb
+++ b/lib/benchmark/ips/job.rb
@@ -188,7 +188,7 @@ module Benchmark
       end
 
       def load_held_results
-        return unless @held_path && !File.zero?(@held_path)
+        return unless @held_path && File.exist?(@held_path) && !File.zero?(@held_path)
         require "json"
         @held_results = {}
         JSON.load(IO.read(@held_path)).each do |result|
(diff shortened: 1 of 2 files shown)


nateberkopec:
Cheers!

nateberkopec:
Released as 2.8.4

ivoanjo:
Thanks for the speedy review/release as well! Happy to contribute :blush:

nateberkopec:
Thanks for submitting a bug report as a PR! With tests! It's like Christmas came early!

PR #89, opened 2018-10-24
Report to stream and iron out compare abstraction

kbrock:
This unifies the double `@stdout` and `@suite` calls. It also allows a user to choose a different output.

Addresses #81 for `warmup` and `benchmark` but not `compare!`

Outstanding:

Changing `compare` is not as easy, the output needs to be passed in:
1. Pass in the stream from one of the `@outs`
2. Pass in a stream to `config` and use that throughout
3. Enhance `StdoutReport` and change `compare` to use the `@outs`.

Let me know which you approach you prefer. I can fixup this PR or just make another.

### Sample Code

```ruby
require "benchmark/ips"

my_stream = StringIO.new
Benchmark.ips do |x|
  x.config(:quiet => true, :suite => ::Benchmark::IPS::Job::StdoutReport.new(my_stream))

  x.report("mul") { 2 * 2 * 2 * 2 * 2 * 2 * 2 * 2 }
  x.report("pow") { 2 ** 8 }

  # x.compare!
end

puts "results: #{my_stream.string}"
```

PR #111, opened 2021-03-04
Report to stream

kbrock:
reopening #89 - you can set this up to output to stdout or a string.

Sorry to jump the gun and close this.
The introduction of these extra methods in the interface just never felt quite right to me.
If you have some feedback, I can hopefully update to something that feels better to both of us.


lib/benchmark/ips/job.rb
```diff
@@ -292,8 +291,7 @@ def run_benchmark
         @list.each do |item|
           next if run_single? && @held_results && @held_results.key?(item.label)
 
-          @suite.running item.label, @time if @suite
```
nateberkopec, line 295: All the various cleanups around removing `if @stdout` are really nice.

nateberkopec:
I think this is super solid as-is. 

Evan made a comment in the original PR re: compare!

> As for how to deal with compare!, I’d prefer to delegate it to the report itself.

@kbrock do you have the bandwidth to do that or would you prefer someone else take it from here?

kbrock:
Think I like the way #113 uses a no-op object to avoid the nil checks.
That one makes fewer changes to the interface of those objects. Which is good.

It doesn't address the ambiguous output that ends up on stdout. Thinking after that one is merged, this approach may change.

nateberkopec:
@kbrock Definitely, I too noticed the overlap between the two PRs

nateberkopec:
@kbrock So, where does this leave us after the suite-accessor PR?

kbrock:
I rebased it and have a question around approaches

I like adding the stdout. that is basically done and I can add a spec to test that

I have never been a fan of the double output, the stdout and report output together. But having an output aggregator feels contrived and like I'm trying too hard to use a design pattern. I also don't like adding to the report interface just so I can make the 2 similar.

Any thoughts?

nateberkopec:
@kbrock Take a look at how Minitest manages basically the same problem, maybe it will spark some inspiration:

https://github.com/seattlerb/minitest/blob/3c6576a51f4e266996e3459d7a0dd054eb4c87f7/lib/minitest.rb#L125
https://github.com/seattlerb/minitest/blob/3c6576a51f4e266996e3459d7a0dd054eb4c87f7/lib/minitest.rb#L822

Otherwise, hard to say without seeing more of the implementation you have in mind.

kbrock:
ok, rebased. Gosh, this has been languishing for far too long. Sorry about that.

- The change to stdout report feels like a no brainer.
- Removing the `Noop{Report,Suite}` feels good. and just using a stdout w/ `/dev/null` or something simplifies quite a bit

This does the multiple outputs the same way that Mini-test does it. This has been my sticking point for years, but since it is good enough for mini test, it should be good enough for me.

Let me know if/where you have apprehensions. I'm more than happy to peel anything out of here into a simpler PR if you want.

lib/benchmark/ips/job/stdout_report.rb
```diff
@@ -2,42 +2,43 @@ module Benchmark
   module IPS
     class Job
       class StdoutReport
```
nateberkopec, line 4: Now that this doesn't necessarily print to Stdout, should we call it something different? Like `ConsoleReport` or something? 🤔

lib/benchmark/ips/job/multi_report.rb
```diff
@@ -0,0 +1,67 @@
...
+    class Job
+      class MultiReport
+        # @returns out [Array<StdoutReport>] list of reports to send output
+        attr_accessor :out
+
+        def empty?
+          @out.empty?
+        end
+
+        def quiet?
```
nateberkopec, line 12: This is an odd definition of `quiet` if you think about it. I think I would kind of expect quiet? and empty? to be the same thing? I understand you probably wrote it this way to keep the behavior the same with the previous implementation.

Should quieting the multireporter really only delete StdoutReports? 🤔

nateberkopec (approved):
I'm OK with this as-is but would appreciate responses to my Q's too

lib/benchmark/ips/job/multi_report.rb
kbrock: Agreed. this feels like a very ugly hack.
One that I don't want my name in `git blame` in a year.

I would prefer not to have a bunch of `if` checks.
You have ideas on this one?

kbrock: ok, my interpretation of `quiet` is a way to quiet the output so we can use the suite and not have so much noise.

In bc84a73ac3d9f6d we kinda lost some of the intent there.



```ruby
      if defined? Benchmark::Suite and Suite.current
        suite = Benchmark::Suite.current
      end

      quiet ||= (suite && suite.quiet?)

      job = Job.new({:suite => suite,
                     :quiet => quiet
      })
```

nateberkopec:
Thanks for your work, @kbrock 🙇

kbrock:
@nateberkopec thanks for picking up so many gems and really rocking it.
Let me know if you need anything related to this or tangental.

ioquatix:
Nice work.

Issue #81, opened 2016-11-11
Allow to return the report output to a String instead of printing it on STDOUT


fidalgo:
It would be nice, to have a way to capture the output to a String in order to print it or send it to an external service, like Bugsnag. Also in a PASS like heroku it's hard to get the reports because the logs are combined from logger, stdout and stderr.

By output I mean the full report:

```
Calculating -------------------------------------
            addition    71.254k i/100ms
           addition2    68.658k i/100ms
           addition3    83.079k i/100ms
addition-test-long-label
                        70.129k i/100ms
-------------------------------------------------
            addition     4.955M (± 8.7%) i/s -     24.155M
           addition2    24.011M (± 9.5%) i/s -    114.246M
           addition3    23.958M (±10.1%) i/s -    115.064M
addition-test-long-label
                         5.014M (± 9.1%) i/s -     24.545M

Comparison:
           addition2: 24011974.8 i/s
           addition3: 23958619.8 i/s - 1.00x slower
addition-test-long-label:  5014756.0 i/s - 4.79x slower
            addition:  4955278.9 i/s - 4.85x slower
```

zenspider:
That should be easy to do, IIRC. Wanna take a whack at it?

ddnexus:
Meanwhile you can do something like:

```ruby
def capture_ips_report(label, &block)
  old_stdout = $stdout
  $stdout    = StringIO.new

  ::Benchmark.ips do |x|
    # add more stuff here
    x.report(label) do
      block.call
    end
  end

  # return the captured output
  $stdout.string
ensure
  $stdout = old_stdout
end
```

ioquatix:
The function that does the printing should just take an optional output stream. It's the simplest and most flexible solution - can output to pipe, stderr, stringio, etc

kbrock:
if you don't want the comparison results, you create a suite and have that output to the appropriate output.

`capture!` will not be captured, that may be possible at some time.

```ruby
my_stream = StringIO.new
Benchmark.ips(:quiet => true, :suite => MySuite.new(my_stream)) do |x|
  x.report(label) { block.call }
end
```

`MySuite` can probably be ripped off of [StdoutReport](https://github.com/evanphx/benchmark-ips/blob/master/lib/benchmark/ips/job/stdout_report.rb)

I'll see if I can tweak that file to let you output to any file.
The changes to `Compare` will probably be more work.

kbrock:
Does #89 meet your needs?
[...]

PR #113, opened 2021-03-17
Allow Suite To be Set via Accessor

### LINKED ISSUE #112: Suite Only Availabe via config
This isn't possible today:

```ruby
Benchmark.ips do |x|
  x.suite = suite
  x.warmup = 3
  x.time = 10
  x.iterations = 5
end
```

but this is

```ruby
Benchmark.ips do |x|
  x.config(
    suite: suite,
    warmup: 3,
    time: 10,
    iterations: 5,
  )
```

Is not being able to set `x.suite = ` a miss?  If so, I'm happy to PR that change.

allcentury:
This aims to close #112.

I removed references to `Benchmark::Suite` - I couldn't find anywhere
this was used and the original commit goes back a long ways.  This looks
like a relic to a class or module that I'm not able to find in git
history via

```
git grep 'Suite' $(git rev-list --all)
```

All config gets passed into `job.config job_opts` and now the block is
evaluated last allowing overrides.

### DIFF
--- a/lib/benchmark/ips/job.rb
+++ b/lib/benchmark/ips/job.rb
@@ -51,16 +51,20 @@ module Benchmark
       # @return [Integer]
       attr_accessor :confidence
 
+      # Silence output
+      # @return [Boolean]
+      attr_reader :quiet
+
+      # Suite
+      # @return [Benchmark::IPS::NoopSuite]
+      attr_reader :suite
+
       # Instantiate the Benchmark::IPS::Job.
-      # @option opts [Benchmark::Suite] (nil) :suite Specify Benchmark::Suite.
-      # @option opts [Boolean] (false) :quiet Suppress the printing of information.
       def initialize opts={}
-        @suite = opts[:suite] || nil
-        @stdout = opts[:quiet] ? nil : StdoutReport.new
         @list = []
-        @compare = false
         @run_single = false
         @json_path = false
+        @compare = false
         @held_path = nil
         @held_results = nil
 
@@ -88,6 +92,20 @@ module Benchmark
         @iterations = opts[:iterations] if opts[:iterations]
         @stats = opts[:stats] if opts[:stats]
         @confidence = opts[:confidence] if opts[:confidence]
+        self.quiet = opts[:quiet]
+        self.suite = opts[:suite]
+      end
+
+      def quiet=(val)
+        @stdout = reporter(quiet: val)
+      end
+
+      def suite=(suite)
+        @suite = suite || Benchmark::IPS::NoopSuite.new
+      end
+
+      def reporter(quiet:)
+        quiet ? NoopReport.new : StdoutReport.new
       end
 
       # Return true if job needs to be compared.
@@ -223,19 +241,19 @@ module Benchmark
 
       def run
         if @warmup && @warmup != 0 then
-          @stdout.start_warming if @stdout
+          @stdout.start_warming
           @iterations.times do
             run_warmup
           end
         end
 
-        @stdout.start_running if @stdout
+        @stdout.start_running
 
         @iterations.times do |n|
           run_benchmark
         end
 
-        @stdout.footer if @stdout
+        @stdout.footer
       end
 
       # Run warmup.
@@ -243,8 +261,8 @@ module Benchmark
         @list.each do |item|
           next if run_single? && @held_results && @held_results.key?(item.label)
 
-          @suite.warming item.label, @warmup if @suite
-          @stdout.warming item.label, @warmup if @stdout
+          @suite.warming item.label, @warmup
+          @stdout.warming item.label, @warmup
 
           Timing.clean_env
 
@@ -280,8 +298,8 @@ module Benchmark
             item.call_times cycles
           end
 
-          @stdout.warmup_stats warmup_time_us, @timing[item] if @stdout
-          @suite.warmup_stats warmup_time_us, @timing[item] if @suite
+          @stdout.warmup_stats warmup_time_us, @timing[item]
+          @suite.warmup_stats warmup_time_us, @timing[item]
 
           break if run_single?
         end
@@ -292,8 +310,8 @@ module Benchmark
         @list.each do |item|
           next if run_single? && @held_results && @held_results.key?(item.label)
 
-          @suite.running item.label, @time if @suite
-          @stdout.running item.label, @time if @stdout
+          @suite.running item.label, @time
+          @stdout.running item.label, @time
 
           Timing.clean_env
 
@@ -334,8 +352,8 @@ module Benchmark
             rep.show_total_time!
           end
 
-          @stdout.add_report rep, caller(1).first if @stdout
-          @suite.add_report rep, caller(1).first if @suite
+          @stdout.add_report rep, caller(1).first
+          @suite.add_report rep, caller(1).first
 
           break if run_single?
         end
--- a/lib/benchmark/ips.rb
+++ b/lib/benchmark/ips.rb
@@ -5,8 +5,10 @@ require 'benchmark/ips/stats/stats_metric'
 require 'benchmark/ips/stats/sd'
 require 'benchmark/ips/stats/bootstrap'
 require 'benchmark/ips/report'
+require 'benchmark/ips/noop_suite'
 require 'benchmark/ips/job/entry'
 require 'benchmark/ips/job/stdout_report'
+require 'benchmark/ips/job/noop_report'
 require 'benchmark/ips/job'
 
 # Performance benchmarking library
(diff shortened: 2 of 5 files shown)


lib/benchmark/ips/job.rb
```diff
@@ -88,6 +92,28 @@ def config opts
         @iterations = opts[:iterations] if opts[:iterations]
         @stats = opts[:stats] if opts[:stats]
         @confidence = opts[:confidence] if opts[:confidence]
+        self.quiet = opts[:quiet]
```
kbrock, line 95: do you want to put an `if opts[:quiet]` guard clause here and with `:suite`?

lib/benchmark/ips/job.rb
```diff
@@ -88,6 +92,28 @@ def config opts
...
+      def quiet=(val)
+        set_reporter(quiet: val)
+      end
+
+      def suite=(suite)
+        if suite
+          @suite = suite
+        else
+          @suite = Benchmark::IPS::NoopSuite.new
+        end
```
kbrock, line 108: is it possible to simplify this?
```suggestion
        @suite = suite || Benchmark::IPS::NoopSuite.new
```

lib/benchmark/ips/job.rb
```diff
@@ -223,28 +249,28 @@ def clear_held_results
 
       def run
         if @warmup && @warmup != 0 then
-          @stdout.start_warming if @stdout
+          @stdout.start_warming
```
kbrock, line 252: thanks - this is much nicer

lib/benchmark/ips/job.rb
```diff
@@ -88,6 +92,28 @@ def config opts
...
+
+      def suite=(suite)
+        if suite
+          @suite = suite
+        else
+          @suite = Benchmark::IPS::NoopSuite.new
+        end
+      end
+
+      def set_reporter(quiet:)
```
kbrock, line 111: since this only accepts quiet, does it make sense to just move the contents of this over to `def quiet=`

or do you have plans on allowing a report to be passed into here in the future?
If you wanted that, you could do `quiet=true` or `reporter=NoopReport.new`

lib/benchmark/ips/job.rb
```diff
@@ -51,16 +51,20 @@ class Job
       # @return [Integer]
       attr_accessor :confidence
 
+      # Silence output
+      # @return [Boolean]
+      attr_reader :quiet
```
kbrock, line 56: does this need to be stored? does it make sense just comparing with noop class?
Is reading `quiet` even used / is it necessary?

kbrock (approved):
this looks nice.

not a fan of the `set_reporter` method but the removal of the if statements sure makes this nicer.

lib/benchmark/ips/job.rb
allcentury: I have a generalist point of view that any state that can be publicly set should be get-able.  `quiet` is part of the public API in a few ways:

```ruby
Benchmark.ips(quiet: true) { }
```

or 

```ruby
Benchmark.ips do |x|
  x.quiet = true
end
```

lib/benchmark/ips/job.rb
allcentury: I've pushed the conditional to `quiet=` at this time so no.

lib/benchmark/ips/job.rb
allcentury: I like this - thanks!

lib/benchmark/ips/job.rb
allcentury: Great points.  I was thinking `reporter` should eventually use dependency injection and if no report is passed in then we'll default to `Noop`.  Then, we can actually decommission `quiet` from the public API and instead, if you don't want output you pass in the `NoopReport`.  I think the idea of a `reporter` and `quiet` are conflating the same thing but getting rid of `quiet` seemed like a breaking change at this time.

allcentury (commented):
Thanks for the review @kbrock !

lib/benchmark/ips/job.rb
allcentury: Maybe the answer is to have a public `reporter=` method that `self.quiet` calls and eventually we can deprecate it?  thoughts on that?

kbrock: I like that idea a lot. it is consistent with your current approach and the approach this project seems to be headed.
At one time, quiet made sense. But now that people want more control over capturing the output, introducing a reporter accessor sounds good.

lib/benchmark/ips/job.rb
kbrock: I was worried that this would overwrite values, but since this is in the initializer, we actually very much do not want the conditional guard clause here. b/c we always want to set this to one of the 2 reporters.

kbrock (commented):
This is looking nice.
I like where you are going with `reporter=`

But please note, that while I have contributed a few PRs here, I am not the maintainer so take my words with a grain of salt.

I have put in a few PRs around better capturing the output, that is why I'm vocal here. A number of people have expressed the desire to have better output / hence another pr that I have open. this seems to overlap that one a bit - so keep it coming and I'll refactor that one.

nateberkopec:
Hey @allcentury - do you have more to add here or are you waiting on my input?

allcentury:
@kbrock - yes, I follow the [`fixup`](https://thoughtbot.com/blog/autosquashing-git-commits) flow until it's ready to merge, then I squash.  

@nateberkopec - thanks!  Ya, could use your input:

option A:

1. deprecate `quiet` in favor of setting a `reporter`.  
2. If you want no output, use the `NoopReporter`.  
3. Replace docs on `quiet` with `NoopReporter`
4. Eventually let anyone inject their own `Reporter` as long as it confirms to the [interface](https://github.com/evanphx/benchmark-ips/blob/master/lib/benchmark/ips/job/stdout_report.rb)
5. Do we merge format + reporter?  `CsvReporter`, `JsonReporter`, etc. 

option B:

1. Keep `quiet` and don't surface the `reporter` idea
2. Nothing to deprecate/change
3. Don't merge format config with the output config

Initially I thought `reporter` could allow users to write their own formatters like `csv`, `json`, etc.  However I see there is a [format option](https://github.com/evanphx/benchmark-ips/blob/5f9a28420688fb6c909af31d950724ab496f1bf2/lib/benchmark/ips.rb#L84-L90).  As it stands w/ this PR:

`StdoutReport`:

1.  Sends everything to stdout 

`NoopReport`:

1. No output

`:format` 

1. Raw or Human, used by the Report classes above.

Can you think of a good reason to have users inject their own `Report` class?  I can't without conflating format.  So I'm leaning towards Option B now.  Thoughts?

kbrock:
> Can you think of a good reason to have users inject their own Report class? I can't without conflating format. So I'm leaning towards Option B now. Thoughts?

There was a request https://github.com/evanphx/benchmark-ips/issues/81 to be able to capture the output. So as long as you could set a string output buffer or something, then that would move this gem towards solving that issue. I had started on a PR to solve that issue but I like where this PR is going and can fix the other PR after you are done here.

nateberkopec:
Well, I don't have any desire to get rid of `quiet`. I think setting `quiet` should just silently swap out the reporter in the background. So I guess that makes me option B? I don't think the other decisions/points in the PR need to be solved today, maybe we merge reporters/formatters in the future but I don't think that has to be solved now.

allcentury:
@nateberkopec - thanks for that.  I think this is ready for your review then.

lib/benchmark/ips/job.rb
```diff
@@ -88,6 +92,24 @@ def config opts
...
+
+      def quiet=(val)
+        set_reporter(quiet: val)
+      end
+
+      def suite=(suite)
+        @suite = suite || Benchmark::IPS::NoopSuite.new
+      end
+
+      def set_reporter(quiet:)
```
kbrock, line 107: I think this is the only part that is suspect.

introducing report= is ok - or not. Just use that and set the value based upon quiet.
but the set_reporter doesn't seem right to me

kbrock (commented):
I like all but that one method.
since it is only used in the one spot, this seems a simple change

but obviously, I defer to nick on this all (since he is supporting the gem after all)

lib/benchmark/ips/job.rb
nateberkopec: @allcentury Purely from a style perspective, `get_` and `set_` methods don't belong in Ruby for me. It looks like PHP or JS. I think this should just be `reporter=`. I have no problem with what the method does.

nateberkopec (changes requested):

nateberkopec:
1 rename and it's GTG

allcentury:
@nateberkopec perfect, thank you!  I removed the `set` and now `reporter` is a pure function which `quiet=` relies on.

nateberkopec:
Thank _you_ @allcentury

kbrock:
thank you @allcentury

Issue #116, opened 2021-05-23
`cannot load such file -- benchmark/ips/noop_suite`


ioquatix:
~~~
Failure/Error: require 'benchmark/ips'

LoadError:
  cannot load such file -- benchmark/ips/noop_suite
# /Users/samuel/.gem/ruby/3.0.0/gems/benchmark-ips-2.9.0/lib/benchmark/ips.rb:8:in `require'
# /Users/samuel/.gem/ruby/3.0.0/gems/benchmark-ips-2.9.0/lib/benchmark/ips.rb:8:in `<top (required)>'
# ./spec/db/benchmark_spec.rb:23:in `require'
# ./spec/db/benchmark_spec.rb:23:in `<top (required)>'
~~~

I'm not sure what's going on, but the latest release seems broken.

I'd suggest you change to use `require_relative` everywhere too: https://github.com/evanphx/benchmark-ips/blob/dea7eb1df656f40c2455a4d69ede8601ba82bf5f/lib/benchmark/ips.rb#L8

~~~
gem "benchmark-ips", "~> 2.8.0"
~~~

fixes the issue for me.

okeeblow:
https://github.com/evanphx/benchmark-ips/pull/115

evanphx:
2.9.1 is out with the fix, sorry about that!

ioquatix:
Thanks!

PR #115, opened 2021-05-22
Fix missing `noop_suite` and `job/noop_report` files in v2.9.0 Manifest

okeeblow:
Previously: https://github.com/evanphx/benchmark-ips/issues/103
Previously previously: https://github.com/evanphx/benchmark-ips/issues/74
Previously previously previously: https://github.com/evanphx/benchmark-ips/issues/59

Before:

```[gst-master] [okeeblow@emi#CHECKING-YOU-OUT] ./bin/benchmark
Traceback (most recent call last):
        2: from ./bin/benchmark:7:in `<main>'
        1: from /usr/local/lib/site_ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require'
/usr/local/lib/site_ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require': cannot load such file -- benchmark/ips (LoadError)
        6: from ./bin/benchmark:7:in `<main>'
        5: from /usr/local/lib/site_ruby/2.7.0/rubygems/core_ext/kernel_require.rb:156:in `require'
        4: from /usr/local/lib/site_ruby/2.7.0/rubygems/core_ext/kernel_require.rb:168:in `rescue in require'
        3: from /usr/local/lib/site_ruby/2.7.0/rubygems/core_ext/kernel_require.rb:168:in `require'
        2: from /home/okeeblow/.gems/gems/benchmark-ips-2.9.0/lib/benchmark/ips.rb:8:in `<top (required)>'
        1: from /usr/local/lib/site_ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require'
/usr/local/lib/site_ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require': cannot load such file -- benchmark/ips/noop_suite (LoadError)
```

Fix:

```[gst-master] [okeeblow@emi#CHECKING-YOU-OUT] curl -o /home/okeeblow/.gems/gems/benchmark-ips-2.9.0/lib/benchmark/ips/noop_suite.rb https://raw.githubusercontent.com/evanphx/benchmark-ips/master/lib/benchmark/ips/noop_suite.rb
[gst-master] [okeeblow@emi#CHECKING-YOU-OUT] curl -o /home/okeeblow/.gems/gems/benchmark-ips-2.9.0/lib/benchmark/ips/job/noop_report.rb https://raw.githubusercontent.com/evanphx/benchmark-ips/master/lib/benchmark/ips/job/noop_report.rb
```

After:

[...]

PR #117, opened 2021-05-24
Include all *.rb files in lib/ into the gem

janko:
Alternative fix for https://github.com/evanphx/benchmark-ips/issues/116. This is the strategy I use in all of my gems, to include all ruby files in the `lib/` directory. I also removed `Manifest.txt`, so that it doesn't need to be updated anymore.


evanphx:
I'd rather manually maintain the list.

eregon:
Unfortunate, since maintaining that list manually has proven to be error-prone and caused already 4 broken releases: https://github.com/evanphx/benchmark-ips/pull/115#issue-650569811

But here is another way to ensure this issue does not happen again: https://github.com/evanphx/benchmark-ips/pull/118

PR #118, opened 2021-06-12
Add test to ensure .rb files under lib are included in the manifest

eregon:
So then issues like https://github.com/evanphx/benchmark-ips/pull/115 which caused 4 broken releases should not happen again.
Alternative to https://github.com/evanphx/benchmark-ips/pull/117

### DIFF
--- a/test/test_benchmark_ips.rb
+++ b/test/test_benchmark_ips.rb
@@ -13,6 +13,13 @@ class TestBenchmarkIPS < Minitest::Test
     $stdout = @old_stdout
   end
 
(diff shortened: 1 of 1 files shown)


evanphx:
Thank you!

test/test_benchmark_ips.rb
```diff
@@ -13,6 +13,13 @@ def teardown
     $stdout = @old_stdout
   end
 
+  def test_all_lib_files_are_included_in_manifest
```
nateberkopec, line 16: Thanks @eregon!

PR #119, opened 2021-10-07
! Fix quiet mode broken when config called without key `quiet`

PikachuEXE:
I have built some stuff relying on quiet mode several year ago
I recently try to run it again and found that it's broken due to this gem outputting results to stdout unexpectedly

Thus I found out about this bug


PikachuEXE:
@nateberkopec 
That's quick thx!
Even better if I can see this gets released soon :D

Issue #120, opened 2022-01-06
Paper cut: `FloatDomainError: Infinity` if warmup period is "small enough"


ivoanjo:
Hello there, and as always, thanks for the amazing job on benchmark-ips!

I noticed this paper cut on one of the tests we do on [Datadog's ddtrace gem](https://github.com/DataDog/dd-trace-rb): if the warm up time is "small enough", we can trigger a division by `0.0`.

Here's an extreme example:

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  gem 'benchmark-ips'
end

Benchmark.ips do |x|
  x.config(:warmup => 0.0000000001, :time => 1)
  x.report("addition") { 1 + 2 }
  x.compare!
end
```

And the result:

```
$ ruby bench-ips-testcase.rb
Warming up --------------------------------------
            additionTraceback (most recent call last):
	9: from bench-ips-testcase.rb:9:in `<main>'
	8: from /Users/ivo.anjo/.rvm/gems/ruby-2.7.5/gems/benchmark-ips-2.9.2/lib/benchmark/ips.rb:53:in `ips'
	7: from /Users/ivo.anjo/.rvm/gems/ruby-2.7.5/gems/benchmark-ips-2.9.2/lib/benchmark/ips/job.rb:247:in `run'
	6: from /Users/ivo.anjo/.rvm/gems/ruby-2.7.5/gems/benchmark-ips-2.9.2/lib/benchmark/ips/job.rb:247:in `times'
	5: from /Users/ivo.anjo/.rvm/gems/ruby-2.7.5/gems/benchmark-ips-2.9.2/lib/benchmark/ips/job.rb:248:in `block in run'
	4: from /Users/ivo.anjo/.rvm/gems/ruby-2.7.5/gems/benchmark-ips-2.9.2/lib/benchmark/ips/job.rb:263:in `run_warmup'
	3: from /Users/ivo.anjo/.rvm/gems/ruby-2.7.5/gems/benchmark-ips-2.9.2/lib/benchmark/ips/job.rb:263:in `each'
	2: from /Users/ivo.anjo/.rvm/gems/ruby-2.7.5/gems/benchmark-ips-2.9.2/lib/benchmark/ips/job.rb:294:in `block in run_warmup'
[...]

PR #121, opened 2022-01-17
Ensure to always run at least 1 iteration for both warmup and for measurement

eregon:
* Fixes https://github.com/evanphx/benchmark-ips/issues/120
* Using do-while loops to ensure the loops run at least once.

Also: Reduce the time needed to runs the tests
* From about 80s to 35s.

cc @ivoanjo @nateberkopec


eregon:
I moved the CI to GitHub Actions since TravisCI no longer works.

ruby-head fails probably due to too old rake, I'll just drop it for now to not add too many changes in this PR: https://github.com/eregon/benchmark-ips/runs/4840536252?check_suite_focus=true
And I test on Ubuntu only  since it's fairly slow on macOS (TravisCI also tested only on Ubuntu).

eregon:
Passing run: https://github.com/eregon/benchmark-ips/actions/runs/1707890584

nateberkopec:
> I moved the CI to GitHub Actions since TravisCI no longer works.

Wow, thank you so much for that!

PR #122, opened 2022-01-25
Remove Hoe, 2.9.3

nateberkopec:


eregon:
Nice, `hoe` always felt more like a hurdle than a help to me :)

Regarding `test_all_lib_files_are_included_in_manifest` you can just remove it now that a `git ls-files` (or a `Dir.glob`) is used (it was ensuring the Manifest.txt was up to date).

nateberkopec:
> Nice, hoe always felt more like a hurdle than a help to me :)

Yup, just a bit of a relic of the pre-bundler era I think.

kbrock:
wow. blast from the past. good job

PR #123, opened 2022-02-09
`compare!(order: :baseline)` option

casperisfine:
This is just a quick proof of concept to gather some feedback, I totally realize it's likely not desirable to change the existing behavior, but I'm willing to do the work to put this behind an option of some sort.

### Use case.

Most of the time when I use `benchmark-ips`, it looks like this:

```ruby
Benchmark.ips do |x|
  x.report("original") {}

  x.report("opt_1") {}
  x.report("opt_2") {}
  # ...
  x.compare!
end
```

Giving a result like:

```
             original    96.197k i/100ms
             opt_1       63.076k i/100ms
             opt_2      250.949k i/100ms
             original 953.227k (± 0.9%) i/s -      4.810M in   5.046288s
             opt_1    653.881k (± 5.0%) i/s -      3.280M in   5.028459s
             opt_2      2.494M (± 0.3%) i/s -     12.547M in   5.031949s

Comparison:
             opt_2:     2493580.5 i/s
             original:   953227.3 i/s - 2.62x  (± 0.00) slower
             opt_1:      653881.1 i/s - 3.81x  (± 0.00) slower
```

But maybe it's just me, I find this output a bit confusing, because what I want to know is how much slower/faster each optimization is compare to "the baseline", which generally is the first registered `x.report`.

So something like this:

```
             original    93.207k i/100ms
             opt_1       61.370k i/100ms
             opt_2      255.237k i/100ms
             original    933.290k (± 0.8%) i/s -      4.754M in   5.093679s
             opt_1       642.273k (± 4.0%) i/s -      3.253M in   5.072394s
             opt_2         2.538M (± 0.2%) i/s -     12.762M in   5.029163s

Comparison:
             original:   933290.0 i/s
             opt_2:     2537580.1 i/s - 2.72x  (± 0.00) faster
             opt_1:      642272.7 i/s - 1.45x  (± 0.00) slower
```

Which is pretty much what I implemented in a quick and dirty way.

### Possible API

So if this feature were to be accepted, what would be the syntax for it?

I'm thinking that it could be mark by using `x.baseline` instead of `x.report` ?:

```ruby
Benchmark.ips do |x|
  x.baseline("original") {}

  x.report("opt_1") {}
  x.report("opt_2") {}
  # ...
  x.compare!
end
```

### DIFF
--- a/lib/benchmark/compare.rb
+++ b/lib/benchmark/compare.rb
@@ -26,31 +26,67 @@ module Benchmark
   #   Reduce using to_proc:   247295.4 i/s - 1.13x slower
   #
   # Besides regular Calculating report, this will also indicates which one is slower.
+  #
+  # +x.compare!+ also takes an +order: :baseline+ option.
+  #
+  # Example:
+  #  > Benchmark.ips do |x|
+  #   x.report('Reduce using block')   { [*1..10].reduce { |sum, n| sum + n } }
+  #   x.report('Reduce using tag')     { [*1..10].reduce(:+) }
+  #   x.report('Reduce using to_proc') { [*1..10].reduce(&:+) }
+  #   x.compare!(order: :baseline)
+  # end
+  #
+  # Calculating -------------------------------------
+  #   Reduce using block    886.202k (± 2.2%) i/s -      4.521M in   5.103774s
+  #     Reduce using tag      1.821M (± 1.6%) i/s -      9.111M in   5.004183s
+  # Reduce using to_proc    895.948k (± 1.6%) i/s -      4.528M in   5.055368s
+  #
+  # Comparison:
+  #   Reduce using block:   886202.5 i/s
+  #     Reduce using tag:  1821055.0 i/s - 2.05x  (± 0.00) faster
+  # Reduce using to_proc:   895948.1 i/s - same-ish: difference falls within error
+  #
+  # The first report is considered the baseline against which other reports are compared.
   module Compare
 
     # Compare between reports, prints out facts of each report:
     # runtime, comparative speed difference.
     # @param entries [Array<Report::Entry>] Reports to compare.
-    def compare(*entries)
+    def compare(*entries, order: :fastest)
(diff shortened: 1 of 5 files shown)


kbrock:
neat idea.

Just having a baseline option (where we assume the first is the baseline?) may be minimal changes.

you would just need:

```ruby
      best = entries.shift if baseline?
      sorted = entries.sort_by{ |e| e.stats.central_tendency }.reverse
      best ||= sorted.shift
```

and the changes to handle `faster`/`slower`

nateberkopec:
Yeah I feel like most of the time you just put the baseline first right?

Seems like an option like `x.compare!(order: :baseline:)` would work?

casperisfine:
Done. 

I couldn't really see how to test it however, as the current test don't really cover the rendering.

lib/benchmark/compare.rb
```diff
@@ -26,31 +26,67 @@ module Benchmark
...
+  # Reduce using to_proc:   895948.1 i/s - same-ish: difference falls within error
+  #
+  # The first report is considered the baseline against which other reports are compared.
   module Compare
 
     # Compare between reports, prints out facts of each report:
     # runtime, comparative speed difference.
     # @param entries [Array<Report::Entry>] Reports to compare.
-    def compare(*entries)
+    def compare(*entries, order:)
```
evanphx, line 56: Order should default to :fastest

evanphx (changes requested):

casperisfine:
@evanphx done.

kbrock:
> Yeah I feel like most of the time you just put the baseline first right?
> 
> Seems like an option like `x.compare!(order: :baseline)` would work?

ooh. I think I spoke too soon. So you do not introduce `baseline` block or variable at all, just `order`. That is potentially nice.

casperisfine:
Thanks!

nateberkopec:
Release coming today or tomorrow

PR #125, opened 2023-02-06
Feat: Support JSON output to STDOUT

okuramasafumi:
I want to support `benchmark-ips` in github-action-benchmark.
https://github.com/benchmark-action/github-action-benchmark

When it can write the result to STDOUT in JSON format, it would be easy to support it.
This commits changes JSON output so that it can directly write to a given argument if it responds to `write`.
It also adds tests and a section about JSON in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,6 +237,29 @@ Benchmark.ips do |x|
 end
 ```
 
+### Output as JSON
+
+You can generate output in JSON. If you want to write JSON to a file, pass filename to `json!` method:
+
+```ruby
+Benchmark.ips do |x|
+  x.report("some report") {  }
+  x.json! 'filename.json'
+end
+```
+
+If you want to write JSON to STDOUT, pass `STDOUT` to `json!` method and set `quiet = true` before `json!`:
+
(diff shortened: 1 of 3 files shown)


eregon:
I think this is fine, although I would like someone else to review.

One note is becnhmarking on GitHub Actions and shared machines is probably most of the times not a good idea, because it's quite noisy. From that project's README, this seems very noisy indeed: https://benchmark-action.github.io/github-action-benchmark/dev/bench/

evanphx (approved):

kbrock (approved):
This looks nice.

Also, this feels like a ~~big~~ bug fix.
Not sure that this distinction matters, but just putting that out there.

Thank you @okuramasafumi

okuramasafumi:
@eregon Although my primary motivation is to add `github-action-benchmark`, enabling JSON output to STDOUT is still useful to communicate to other tools.

nateberkopec:
Was busy with Puma last week, I'll get to releasing this now :)

大倉さんありがとうございます！

PR #128, opened 2023-03-30
reduces calc overhead of suffix helper method

jtbg:
_if only there were some sort of tool we could use to measure the performance gains_

### DIFF
--- a/lib/benchmark/timing.rb
+++ b/lib/benchmark/timing.rb
@@ -1,5 +1,5 @@
 module Benchmark
-  # Perform caclulations on Timing results.
+  # Perform calculations on Timing results.
   module Timing
     # Microseconds per second.
     MICROSECONDS_PER_SECOND = 1_000_000
--- a/lib/benchmark/ips.rb
+++ b/lib/benchmark/ips.rb
@@ -82,20 +82,15 @@ module Benchmark
     end
 
     module Helpers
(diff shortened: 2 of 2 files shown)


kbrock (approved):
this is nice.

could you squash your commits?
(disclaimer: I'm not a maintainer)

kbrock:
Aah, I was confused because you commented on a commit that had changes not here.
This PR is just picking off the obvious win. 👍

kbrock:
hmm... could you remove the merge commit, and rebase this onto master?

(sorry if you already know this stuff)

```
git fetch upstream
git checkout scale_helper_optimization
git branch scale_helper_optimization_bak # just in case
git reset --hard HEAD~1 # drop the merge commit
git rebase upstream/master

git diff scale_helper_optimization_bak # should be the same
git log -3 # this branch should just have a single commit
```

jtbg:
WHat does `git branch scale_helper_optimization_bak` do here? I've never seen that before

jtbg:
This is why I always get @emgreen33 to untangle my git spaghetti for me :rofl: 
Thanks for the help!

lib/benchmark/timing.rb
```diff
@@ -1,5 +1,5 @@
 module Benchmark
-  # Perform caclulations on Timing results.
```
nateberkopec, line 2: I did it for the lulz!

PR #130, opened 2023-10-27
Automatically print the Ruby description at the top of the report

casperisfine:
It's something I usually manually do when sharing benchmark, but I generally don't see it in most of the benchmark results being shared out there, sometimes making it harder to reproduce the results.

I think it would be nice if benchmark-ips just always printed it.

### DIFF
--- a/lib/benchmark/ips/job/stream_report.rb
+++ b/lib/benchmark/ips/job/stream_report.rb
@@ -8,6 +8,7 @@ module Benchmark
         end
 
         def start_warming
+          @out.puts RUBY_DESCRIPTION
           @out.puts "Warming up --------------------------------------"
         end
 


eregon (approved):
Oh yeah that's a great idea, and I typically have the same problem.

casperisfine:
@nateberkopec @evanphx could I pique your interest in this PR?

zenspider:
POKE

casperisfine:
AFAIK Benoit doesn't have merge permissions, nor is a gem owner on rubygems. Only you or Nate can.

eregon:
I'm happy to become a benchmark-ips maintainer if it helps.
@evanphx If that helps, could you give me both GitHub & RubyGems.org permissions (my email is eregontp on gmail)?

evanphx:
Oh! Ha, my mistake! @eregon has been so active I forgot!

evanphx:
It's done and released!

PR #134, opened 2024-01-11
Provide Benchmark.quick_compare to quickly compare methods on an object

### LINKED ISSUE #133: Add Q&D driver method
I have to look up Benchmark.ips EVERY time and the (ri) doco is [still a PITA](https://github.com/evanphx/benchmark-ips/issues/71#issuecomment-1878166929).

Pls consider a simple top level driver method like so:

```ruby
def quick_bench_ips(*methods, warm:nil, time:nil, show:nil)
  Benchmark.ips do |x|
    x.warmup = warm if warm
    x.time   = time if time
    x.compare!

    if show then
      max = methods.map(&:size).max
      methods.each do |name|
        val = send name
        puts "%*s: %p" % [max, name, val]
      end
      return
    end

    methods.each do |name|
      x.report(name) do |x|
        x.times { send name }
      end
    end
  end
end



def report_tally1
  # ...
end

def report_tally2
  # ...
end

def report_tally3
  # ...
end

quick_bench_ips(:report_tally1,
                :report_tally2,
                :report_tally3)
```

evanphx:
Closes #133 

Any feedback welcome, especially from @zenspider


lib/benchmark/ips/quick.rb
```diff
@@ -0,0 +1,32 @@
+module Benchmark
+  module IPS
+    # Quickly compare multiple methods on the same method.
```
eregon, line 3:
```suggestion
    # Quickly compare multiple methods on the same object.
```

lib/benchmark/ips/quick.rb
```diff
@@ -0,0 +1,32 @@
...
+    # available:
+    #
+    # @param warmup How many seconds to warm up the benchmark process
+    # @param time How many seconds to benchmark each method
+    def quick_compare(obj, *methods)
+      if methods.last.kind_of? Hash
+        opts = methods.pop
+      else
+        opts = {}
+      end
```
eregon, line 17:
```suggestion
    def quick_compare(obj, *methods, **opts)
```
Wouldn't this work?

lib/benchmark/ips/quick.rb
```diff
@@ -0,0 +1,32 @@
...
+
+      Benchmark.ips do |x|
+        x.compare!
+        x.warmup = opts[:warmup] if opts[:warmup]
+        x.time = opts[:time] if opts[:time]
+
+        methods.each do |name|
+          x.report(name) do |x|
+            x.times { obj.__send__ name }
+          end
```
eregon, line 27:
```suggestion
          block = eval "proc { obj.#{name} }"
          x.report(name, &block)
```

This avoids the `__send__` overhead in an easy way and it also reuses the existing block form optimization to use a while loop (instead of times which needs another block call per iteration).

lib/benchmark/ips/quick.rb
```diff
@@ -0,0 +1,32 @@
...
+    # Quickly compare multiple methods on the same method.
+    # @param obj The object to call the methods on
+    # @param methods The list of methods to call as Symbols
+    #
+    # Keyword arguments are detected as well, with the following
+    # available:
+    #
+    # @param warmup How many seconds to warm up the benchmark process
+    # @param time How many seconds to benchmark each method
+    def quick_compare(obj, *methods)
```
zenspider, line 12: Please note that in #133 I said _nothing_ about "quickly comparing methods on an object". Please don't make this take an `obj` argument that you send on. benchmark-ips is MUCH more than running microbenchmarks on Array to know that `reduce` is slower.

Some code needs more complex setup. 
Some code is _comparing_ that setup. 
Or anything else. 

This argument limits things and would require me to make a class with my benchmark methods to get around it. THE POINT of this method is to require less structure to get benchmarks up and running and get an easy comparison of _anything_.

```ruby
ary =  [Object.new] * 969
ary += [Assertion.new] * 3
ary += [UnexpectedError.new] * 1
ary += [Skip.new] * 27

RESULTS = ary

def report_orig
  aggregate = RESULTS.group_by { |r| r.class }
  aggregate.default = [] # dumb. group_by should provide this

  [aggregate[Assertion].size, aggregate[UnexpectedError].size, aggregate[Skip].size]
end

def report_orig2
  aggregate = RESULTS.group_by { |r| r.class }.transform_values(&:size)
  aggregate.default = 0

  [aggregate[Assertion], aggregate[UnexpectedError], aggregate[Skip]]
end

def report_tally
  aggregate = RESULTS.map { |r| r.class }
  aggregate = if RUBY_VERSION > "3.1" then
                aggregate.tally(Hash.new 0)
              else
                aggregate.tally.tap { |r| r.default = 0 }
              end

  [aggregate[Assertion], aggregate[UnexpectedError], aggregate[Skip]]
end

def report_tally2
  aggregate = RESULTS.map { |r| r.class }
  aggregate = aggregate.tally.tap { |r| r.default = 0 }

  [aggregate[Assertion], aggregate[UnexpectedError], aggregate[Skip]]
end

def report_tally3
  aggregate = RESULTS.map { |r| r.class }.tally
  aggregate.default = 0

  [aggregate[Assertion], aggregate[UnexpectedError], aggregate[Skip]]
end
```

lib/benchmark/ips/quick.rb
zenspider: Yeah. This reads like you're still supporting 2.6. Please drop support on 2.6.

lib/benchmark/ips/quick.rb
zenspider: Man... you're really focused on that send... Again: Overhead doesn't matter as long as it is evenly/fairly applied.

I was using the block arg form of report because my understanding is that you get better resolution. I assumed that `x` was tuned based on the warmup numbers. Is this not the case? My benchmark benchmarks show a 20% increase in performance using `x` + `x.times`.

zenspider (commented):
This is not the quick_compare I was hoping for... but not blocking.

lib/benchmark/ips/quick.rb
evanphx: Correct, you didn't say anything about testing on an object. But without a receiver, who do you want the methods called on? just toplevel methods?

lib/benchmark/ips/quick.rb
evanphx: I'd like to still support older versions, folks use benchmark-ips to benchmark across ruby versions.

lib/benchmark/ips/quick.rb
evanphx: I moved `quick_compare` to Kernel so it's more flexible, more along the lines of your original version.

lib/benchmark/ips/quick.rb
eregon: > Again: Overhead doesn't matter as long as it is evenly/fairly applied.

It does. Otherwise some differences cannot be measured with a big enough overhead.
A benchmarking harness should always aim to have minimal overhead.

lib/benchmark/ips/quick.rb
eregon: What's the problem with `Benchmark::IPS.quick_compare(self, :report_orig, :report_orig2, ...)`?
Having the object is useful if you have more state you need to pass in.
Then for instance this state could live in an object instance.

I am against polluting Kernel.
A benchmarking harness should not monkey-patch anything.

Also for 10+ years users of this gem managed just fine without this helper method, so maybe we should not add it.
At least I think we should get multiple opinions to not add something which only works for one person.
The API is also the same as stdlib Benchmark, it's pretty simple.

lib/benchmark/ips/quick.rb
eregon: I think even in 2.6 `**opts` works fine:
```
$ ruby -ve 'def foo(obj, *methods, **options); p methods; p options; end; foo("obj", :bar, a: 1)'
ruby 2.6.9p207 (2021-11-24 revision 67954) [x86_64-linux]
[:bar]
{:a=>1}
```

evanphx: We currently support all the way back to 2.3. We could drop support for older versions, but I usually only do that if doing so makes a material difference. In this case, it's a pretty trivial matter.

zenspider (approved):

nateberkopec:
I'm back from the dead.

1. I too don't like polluting `Kernel` _by default_. I think it's fine if we don't `require 'benchmark/ips/quick` by default. But, then you're kind of missing the point of this whole feature ofc because now you have to add another 15 characters at the top of your benchmark. In my head, the interface that makes the most sense is actually `Benchmark.ips_quick_compare` but I'm neutral and would be fine with `Benchmark::IPS.quick_compare`.
2. I agree that benchmark/ips is about more than microbenchmarking, but microbenchmarking is cool too so we should minimize overhead where it's easy to do so.
3. Supporting 2.3 is rad and doesn't block this feature

If no one objects I'll pick this up and run w/it

lib/benchmark/ips/quick.rb
```diff
@@ -0,0 +1,30 @@
...
+  # Quickly compare multiple methods on the same method.
+  # @param obj The object to call the methods on
+  # @param methods The list of methods to call as Symbols
+  #
+  # Keyword arguments are detected as well, with the following
+  # available:
+  #
+  # @param warmup How many seconds to warm up the benchmark process
+  # @param time How many seconds to benchmark each method
+  def quick_compare(*methods)
```
amomchilov, line 11: Can we use regular kwargs here?

```suggestion
  def quick_compare(*methods, warmup: nil, time: nil)
```

lib/benchmark/ips/quick.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    Benchmark.ips do |x|
+      x.compare!
+      x.warmup = opts[:warmup] if opts[:warmup]
+      x.time = opts[:time] if opts[:time]
+
+      methods.each do |name|
+        x.report(name) do |x|
+          x.times { __send__ name }
```
amomchilov, line 25: How about using one of the faster forms, like the explicit `while` loop, or string form?

nateberkopec:
I've changed it to my take.

1. I tried @eregon's suggestion to send it directly to reduce overhead BUT that means you can't just define a method on main/Kernel and then quick compare it, because that method you just defined is private. The more important part of this feature is convenience rather than suitability for microbenching, so I decided to keep `__send__`
2. It's been moved to `Benchmark.ips_quick`.
3. Expanded docs.

LMK about objections, otherwise I'll add a test and merge

lib/benchmark/ips.rb
```diff
@@ -73,6 +76,34 @@ def ips(*args)
...
+    # @option opts [Integer] :time The number of seconds to run the benchmark. (optional)
+    #
+    # @example Compare String#upcase and String#downcase
+    #   ips_quick([:upcase, :downcase], "hello")
+    #
+    # @example Compare two methods you just defined, with a custom warmup.
+    #   def add; 1+1; end
+    #   def sub; 2-1; end
+    #   ips_quick([:add, :sub], warmup: 10)
+    def ips_quick(methods, receiver = Kernel, **opts)
```
MatheusRich, line 93: I wonder if having quick in the name would make people think that it runs faster than the normal version

MatheusRich (commented):
I love this!

README.md
```diff
@@ -94,6 +94,18 @@ One benefit to using this method is benchmark-ips automatically determines the
...
 guessing iteration counts as we do with the traditional Benchmark library.
 
+You can also use `ips_quick` to save a few lines of code:
+
+```ruby
+Benchmark.ips_quick([:upcase, :downcase], "hello") # runs a suite comparing "hello".upcase and "hello".downcase
+
+def first; MyJob.perform(1); end
+def second; MyJobOptimized.perform(1); end
+Benchmark.ips_quick([:first, second]) # compares :first and :second
```
evanphx, line 104: I think you mean `:second` here.

lib/benchmark/ips.rb
zenspider, line 93: PLEASE consider splatting out `methods`. You can make `receiver` a kwarg.

lib/benchmark/ips.rb
```diff
@@ -73,6 +76,34 @@ def ips(*args)
       report
     end
 
+    # Quickly compare multiple methods on the same object.
+    # @param methods [Array<Symbol>] An array of method names (as symbols) to compare.
+    # @param receiver [Object] The object on which to call the methods. Defaults to Kernel.
+    # @param opts [Hash] Additional options for customizing the benchmark.
+    # @option opts [Integer] :warmup The number of seconds to warm up the benchmark. (optional)
+    # @option opts [Integer] :time The number of seconds to run the benchmark. (optional)
```
amomchilov, line 84: IMO these are really expressed so much more simplify without a kwargs hash.

This is so much more clear, the defaults are explicitly shown, and it doesn't need yarddoc to interpret:

```suggestion
    def ips_quick(methods, receiver = Kernel, warmup: 2, time: 5)
```

nateberkopec:
Thanks for everyone's feedback. This should be the final form now, incl. a basic test. If no objections, will merge in a few days.

lib/benchmark/ips.rb
```diff
@@ -13,11 +13,14 @@
...
-  # @see https://github.com/evanphx/benchmark-ips
+  #
+  # See Benchmark.ips for documentation on using this gem~
+  #
+  # @see {https://github.com/evanphx/benchmark-ips}
   module IPS
 
     # Benchmark-ips Gem version.
-    VERSION = "2.13.0"
+    VERSION = "2.13.1"
```
nateberkopec, line 23: Just saw the version bump here. Changing to 2.14.

PR #140, opened 2024-07-10
Improve code template in README.md

amomchilov:
Some ideas for making the minimal template a bit more readable. See my comments below, would love your thoughts @evanphx

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,11 +27,7 @@ require 'benchmark/ips'
 Benchmark.ips do |x|
   # Configure the number of seconds used during
   # the warmup phase (default 2) and calculation phase (default 5)
-  x.config(:time => 5, :warmup => 2)
-
-  # These parameters can also be configured this way
-  x.time = 5
-  x.warmup = 2
+  x.config(warmup: 2, time: 5)
 
   # Typical mode, runs the block as many times as it can
   x.report("addition") { 1 + 2 }
@@ -43,8 +39,9 @@ Benchmark.ips do |x|
   x.report("addition2") do |times|
     i = 0
     while i < times
-      1 + 2
       i += 1
+
+      1 + 2
     end
   end
 


README.md
```diff
@@ -27,11 +27,7 @@ require 'benchmark/ips'
 Benchmark.ips do |x|
   # Configure the number of seconds used during
   # the warmup phase (default 2) and calculation phase (default 5)
-  x.config(:time => 5, :warmup => 2)
-
-  # These parameters can also be configured this way
-  x.time = 5
-  x.warmup = 2
+  x.config(warmup: 2, time: 5)
```
amomchilov, line 30: It would make sense for `warmup` to come first, to both match the comment, and the runtime order these take place in.

Also, we can migrate off the old-style hash rockets

README.md
```diff
@@ -27,11 +27,7 @@ require 'benchmark/ips'
 Benchmark.ips do |x|
   # Configure the number of seconds used during
   # the warmup phase (default 2) and calculation phase (default 5)
-  x.config(:time => 5, :warmup => 2)
-
-  # These parameters can also be configured this way
```
amomchilov, line 32: IMO these are just redundant. Is there any reason to use them? Perhaps they can live in the API docs, but I don't think they carry their weight in this minimal template.

README.md
```diff
@@ -42,9 +38,8 @@ Benchmark.ips do |x|
   # introduces incorrectable errors.
   x.report("addition2") do |times|
     i = 0
-    while i < times
+    while (i += 1) <= times
```
amomchilov, line 41: Perhaps this is "too clever", but it has the nice effect of not needing benchmark-related code in the loop body. What's left is purely benchmarked code.

eregon: I think it's too error-prone, e.g. `while (i += 1) < times` is very subtly incorrect.
IMO this variant should not be shown in the README or not prominently, `x.report("addition3", "1 + 2")` is simpler if one cares a lot about overhead, there is little value to do the loop manually and it's far more verbose.

amomchilov: > there is little value to do the loop manually

I haven't been able to use the string form much in practice, because it runs in a different context that can't access the local variables you've made outside the eval'ed string.

So I can't use it for anything that requires non-measured setup code, e.g.

```rb
#!/usr/bin/ruby

require "benchmark/ips"

RubyVM::YJIT.enable

s = Set.new.compare_by_identity.freeze

Benchmark.ips do |x|
  # undefined local variable or method `s' for #<Benchmark::IPS::Job::Entry:0x00000001050ffa68> (NameError)
  x.report(".new.compare_by_identity", "s.dup")
  x.report(".dup                    ", "Set.new.compare_by_identity")
  x.compare!
end

```

eregon: One could just use `S = Set.new.compare_by_identity.freeze` for that.
If wanting to measure something small, it's also best to avoid reading local variables from 2 frames up as you would with the manual loop form and this example.

amomchilov: Would reading a constant `S` be faster than a local `s`? 😲

eregon: Yes, it should be, especially given that `s` is two blocks away.
A JIT can even potentially embed the constant in JITed code and there is no read anymore (with safepoints to deopt if it changes).
The interpreter should be able to make it a single read in general.

nateberkopec: If this change is removed or changed to `x.report("addition3", "1 + 2")`, I will merge.

amomchilov: @nateberkopec Done! :)

eregon (approved):
LGTM

amomchilov:
cc @evanphx what do you think of these changes?

nateberkopec:
Thanks for the bump, forgot about this one

PR #146, opened 2025-11-10
Use a json content type to submit the report

arielj:
When sharing the benchmark report to benchmark.fyi, we can set the content type to `application/json` so it's parsed properly by Rails.

For more context, we have to handle the body in a non-standard way manually parsing the request body when the content_type is not specified, something discussed in https://github.com/fastruby/benchmark.fyi/pull/15


nateberkopec:
🙏  ty

PR #148, opened 2026-03-10
Fix Ruby 4.0 warning: use require_relative for internal files

paddor:
Replace all `require` calls for internal files with `require_relative` in `lib/benchmark/ips.rb` and the test suite to fix warnings on Ruby 4.0.

### DIFF
--- a/test/test_benchmark_ips.rb
+++ b/test/test_benchmark_ips.rb
@@ -1,5 +1,5 @@
-require_relative "../lib/benchmark/ips"
 require "minitest/autorun"
+require "benchmark/ips"
 require "stringio"
 require "tmpdir"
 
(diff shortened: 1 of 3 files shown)


eregon:
Which warnings?

eregon:
```
$ bundle exec rake
/home/eregon/code/benchmark-ips/test/test_benchmark_ips.rb:2: warning: benchmark/ips is found in benchmark, which is not part of the default gems since Ruby 4.0.0.
```
That seems a clear bug of https://github.com/ruby/ruby/blob/master/lib/bundled_gems.rb.
It's filed as https://bugs.ruby-lang.org/issues/21828, I'll reopen.

test/test_benchmark_ips.rb
```diff
@@ -1,5 +1,5 @@
+require_relative "../lib/benchmark/ips"
```
eregon: The changes in `lib/` look fine and are good for performance but changes in `test/` are bad because they prevent e.g. testing against an installed `benchmark-ips` gem, so this and below should remain as `require "benchmark/ips"`.

paddor:
Updated

eregon (approved):
LGTM

